PR #107, opened 2026-05-18
fix(skills): honor `..` traversal in skills.include patterns

ejdutton:
## Summary

- `vat build`, `vat verify`, and `vat skills validate` previously failed to discover skills when a `skills.include` pattern stepped out of the package via `..` (e.g. `"../../docs/skills/*/SKILL.md"` in monorepos where SKILL.md sources live alongside, not inside, the package). The shared `discoverSkillsFromConfig` helper handed every pattern to a single downward-only crawl rooted at `projectRoot`, so `..` segments could never resolve to a real file. `vat audit` accepted the same config only because it has a separate filesystem-first walker.
- Each include pattern is now split into a literal base + glob remainder via `picomatch.scan`, patterns are grouped by their resolved absolute base, and the crawler runs once per base. User-supplied excludes stay anchored to `projectRoot` so anchored patterns like `docs/private/**` keep their original meaning. A pattern resolving to a nonexistent base now silently produces zero matches (matching audit's tolerance) rather than crashing.

## Test plan

- [x] New integration test `packages/cli/test/integration/skills-include-parent-traversal.integration.test.ts` — 6 cases (inside, outside via `..`, mixed inside+outside, literal include path, user-supplied excludes, nonexistent base). All four `..`-using cases fail on `main`'s code and pass on this branch.
- [x] `bun run validate` — all phases pass (lint, typecheck, duplication check, unit, integration, system).
- [x] No back-compat shim per pre-1.0 policy; semantics for in-package include patterns (no `..`) are preserved end-to-end (covered by existing integration + system suites).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Fixed
+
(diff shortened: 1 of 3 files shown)


jdutton:
### Code review

No issues found. Checked for bugs and CLAUDE.md compliance.

🤖 Generated with [Claude Code](https://claude.ai/code)

<sub>- If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

PR #115, opened 2026-06-03
fix(cli): derive marketplace publish version from staged plugins

### LINKED ISSUE #110: vat claude marketplace publish reports wrong version in CLI banner, commit message, and status payload
## What I observed

`vat claude marketplace publish` prints the **project's root `package.json` version** in three places — CLI banner, the commit message it pushes to the marketplace repo, and the structured status payload — instead of the version of the plugin/marketplace actually being published.

Example illustrating the bug (fabricated marketplace `acme-tools`, fabricated plugin bumped from `0.0.3` → `0.0.4`, in a project where the root `package.json` version is `0.0.1`):

```
$ vat claude marketplace publish
Publishing marketplace "acme-tools" v0.0.1
   ...
   Commit: abc1234 publish v0.0.1
   Pushed to https://github.com/example-org/acme-tools-claude-marketplace.git branch main
   Tagged source repo: acme-tools-v0.0.4
---
status: success
published:
  - marketplace: acme-tools
    version: 0.0.1
```

Three spots say `v0.0.1`. The `Tagged source repo: ...-v0.0.4` line is correct, and the actual `plugin.json` published into the marketplace correctly contains `"version": "0.0.4"`. So **artifacts are right; the label is wrong.**

`git log` on the resulting marketplace repo shows every publish ever made (v0.0.1, v0.0.2, v0.0.3, v0.0.4) carries the identical commit message `publish v0.0.1`. The bug has been present since the first publish and goes unnoticed when the first publish happens to be 0.0.1.

## Root cause

`readProjectVersion(configDir)` in `packages/cli/src/commands/claude/marketplace/publish.ts:110-126` reads the **root `package.json`** of the project hosting the vat config — not the plugin/marketplace version being staged. That string then flows to all three reporting sites:

- CLI banner — `publish.ts:192`:
  ```ts
  logger.info(`Publishing marketplace "${mpName}" v${version}`);
  ```
- Commit message — `publish.ts:206-210` passes the same `version` (via `composeResult.version`, which is just the option passed in at `publish-tree.ts:62`) to `createCommitMessage()` in `git-publish.ts:45-64`:
  ```ts
  const lines = [`publish v${version}`];
  ```
- `PublishResult.version` in the status payload — `publish.ts:233-234`.

So one wrong variable produces three wrong outputs. Meanwhile `composeResult.publishedPlugins` is built independently by reading the staged `marketplace.json` (`publish-tree.ts:85-94`), which is why per-plugin tag pushes (`pluginTagName(name, version)` in `git-publish.ts:147`) get the correct version.

For any project that hosts a vat config but isn't itself a version-bumped npm package (common — many KB-style or content-hosting projects leave the root `package.json` version at `0.0.1` permanently), this means every publish forever reports `v0.0.1`.

## Expected behavior

The CLI banner, commit message, and status `version` should reflect what was actually published. The natural source is the published marketplace itself (the `version` field in the staged `marketplace.json`, or the resolved plugin version when a marketplace contains a single plugin) — i.e. the same data `publishedPlugins` is already reading.

## Suggested fix

Stop using `readProjectVersion(configDir)` as the label and derive the label from staged content instead.

Two viable options:

1. **Marketplace-level version.** If `marketplace.json` carries its own `version`, read that and use it for the banner/commit/status (parallel to how `publishedPlugins` is sourced in `publish-tree.ts:85-94`).
2. **Plugin version when 1:1.** If a marketplace contains exactly one plugin, fall back to that plugin's version. For multi-plugin marketplaces, either show all bumped plugin versions or drop the `v<X>` from the banner/commit and let the plugin tags carry the version (the source of truth).

Sketch (option 2, single-plugin case):

```ts
// publish.ts, after composePublishTree(...)
const labelVersion =
  composeResult.publishedPlugins.length === 1
    ? composeResult.publishedPlugins[0].version
    : version; // or skip the v<X> entirely

logger.info(`Publishing marketplace "${mpName}" v${labelVersion}`);

const commitMessage = createCommitMessage(
  labelVersion,
  composeResult.changelogDelta,
  sourceRepo ? { sourceRepo } : undefined,
);
// ...and use labelVersion in PublishResult.version
```

Either way the fix is one variable in `publishOneMarketplace` — the call sites already exist.

Worth also reconsidering whether `readProjectVersion` should exist at all in this command. The project-level `package.json` version has no relationship to what's being published; it's a coincidence-of-naming bug magnet.

## Impact

Cosmetic — published artifacts are correct (plugin.json content, source-repo tags). But:

- Misleads operators during routine publishes (the banner is the headline line of output).
- Persists historically inaccurate `publish v0.0.1` commit messages on every marketplace repo published by vat — `git log` on the marketplace repo becomes useless for version archaeology.
- Status payload is unreliable for any automation consuming it.

## Repro

Any project with a previously published marketplace where the plugin version has been bumped at least once and the root `package.json` `version` field has not (the common case for content-hosting projects that use vat as a build tool rather than publish themselves as an npm package). Run:

```
vat claude marketplace publish --dry-run
```

CLI banner reports the root `package.json` version, not the staged marketplace/plugin version.

## Environment

- vat: 0.1.38
- macOS (Darwin 25.4.0)

ejdutton:
## Summary

- Fixes #110: `vat claude marketplace publish` reported the project root `package.json` version in the CLI banner, commit message, status YAML, and CHANGELOG section lookup instead of what was actually being published.
- The label version is now derived from the staged `marketplace.json` — single-plugin marketplaces use the plugin's version (`v0.0.4`); multi-plugin marketplaces drop the `v<X>` entirely since the per-plugin source-repo tags carry the truth.
- Deletes `readProjectVersion`, the misleading helper that caused the bug, and changes `createCommitMessage` to take a `headline` string so the caller owns the rendering decision.

## Visible side-effects (calling out for review)

- **Status YAML's `published[*].version` is now absent for multi-plugin marketplaces.** Previously it carried the misleading project version; now automation should read per-plugin versions from the published `marketplace.json`.
- **Single-plugin marketplaces' stamped CHANGELOG lookup now actually matches.** Previously the `## [X.Y.Z]` lookup used the project version and almost never matched, silently falling back to `## [Unreleased]`. With the fix, a stamped section matching the plugin's version will be used as the commit body.

These are the reasons the CHANGELOG entry sits under `### Changed (breaking, pre-1.0)` rather than `### Fixed` — mirroring PR #114's pattern for the external-URL findings fix.

## Out of scope (deliberate)

- `marketplace.json`'s optional top-level `version` field is not yet consumed (Jeff's Option 1 from the issue). A separate follow-up would extend `ClaudeMarketplaceSchema` so users can declare an aggregate marketplace version in `vibe-agent-toolkit.config.yaml`, then plumb it through the build pipeline.

## Test plan

- [x] New unit tests cover the derivation matrix — `publish-tree.test.ts`
  - 1 plugin → use that plugin's version
  - 2+ plugins → `undefined`
  - no plugin with a usable `version` field → `undefined`
- [x] New unit test for headline-style commit subject (multi-plugin) — `git-publish.test.ts`
- [x] Integration test asserts multi-plugin commit subject is exactly `publish multi-plugin` — `multi-plugin-marketplace.integration.test.ts`
- [x] System test asserts banner + YAML show `v0.4.2` (single-plugin) and never `v1.0.0` (project root) — `marketplace-publish.system.test.ts`
- [x] `bun run validate` passes locally (lint, typecheck, unit, integration, system, duplication)
- [ ] Version bump / RC decision — deferred to you; happy to land as-is under `[Unreleased]` or cut an RC if you'd prefer


jdutton (approved):
**✅ Approved — non-blocking. Nothing here needs to gate merge; land when ready.**

Audited the full diff against `main`. The fix is solid: it identifies and excises the root cause rather than patching around it, the test matrix is good, and the user-visible side-effects are honestly documented in both the PR body and CHANGELOG.

## What it gets right
- **Kills the bug at the source.** `readProjectVersion(configDir)` (read the *project root* `package.json`) is deleted entirely, and the now-unused `readFileSync` import is dropped with it — confirmed it had no other consumer in `publish.ts`.
- **One derivation → four consistent sites.** `labelVersion` flows from the staged `marketplace.json` into the banner, dry-run line, commit headline, and `PublishResult.version`. Old bug was "one wrong variable → three wrong outputs"; this is the inverse.
- **`createCommitMessage(headline, …)`** correctly pushes the render decision to the caller so multi-plugin marketplaces avoid a misleading `v<X>`.
- **`extractChangelogDelta` / `deriveLabelVersion`** extraction is behavior-preserving and lowers complexity in `composePublishTree`.
- **CHANGELOG placement under `### Changed (breaking, pre-1.0)`** (not `### Fixed`) is the right call — the status-YAML shape and commit-message format genuinely change for consumers.
- Verified the "multi-plugin status YAML omits `version`" claim holds: `writeYamlOutput` uses `yaml.stringify`, which drops `undefined` map values.

## Optional follow-ups (none blocking)
1. **`deriveLabelVersion` keys off the *version-filtered* plugin list, not total plugin count.** Edge case: a marketplace with 2 plugins where only 1 carries a `version` yields `length === 1` and gets a single-plugin `v0.x` label. In practice the build pipeline versions every plugin, so this is unlikely — but the intent ("exactly one plugin") and implementation ("exactly one *versioned* plugin") diverge. A comment noting the all-plugins-versioned assumption, or threading the raw count, would close the gap.
2. **Two documented behaviors aren't asserted end-to-end:** the multi-plugin status YAML *omitting* `version` (integration test checks the commit subject instead), and the single-plugin commit subject `publish v0.4.2` (system test is `--dry-run`, so no commit). Both are covered indirectly; a one-line assertion on the multi-plugin YAML absence would lock the contract.

Reviewed by Claude (Opus 4.8) at Jeff's request.

PR #124, opened 2026-06-16
fix(rag): scope lazy-load catches so model failures aren't mislabeled 'not installed' (#118)

### LINKED ISSUE #118: rag: over-broad catch in loadPipeline mislabels all failures as '@xenova/transformers is not installed'
## Problem

`loadPipeline` in `packages/rag/src/embedding-providers/transformers-embedding-provider.ts:34-43` wraps **both** the dynamic import and the model download/inference in a single `try/catch` that always rethrows a fixed "not installed" message:

```ts
try {
  const { pipeline } = await import('@xenova/transformers');   // (A) import
  return (await pipeline('feature-extraction', model, {        // (B) download model + run inference
    quantized: true,
  })) as PipelineFunction;
} catch {
  throw new Error(
    '@xenova/transformers is not installed. Install with: npm install @xenova/transformers',
  );
}
```

Any failure in step (B) — model download failure, network error, `onnxruntime-node` native backend error, OOM, etc. — is reported as **"not installed"**, even when the package is installed and importable. The real error is swallowed (not even attached as `cause`).

## Impact

This actively misleads debugging. A Windows CI failure was initially diagnosed as a missing/optional dependency (and a proposed fix was to move `@xenova/transformers` to `devDependencies`) when the package was in fact installed — the real failure was the model download / native backend on Windows. The test-reliability symptom was fixed by skipping those tests on Windows (see the linked PR), but this misleading error message remains and will mislead the next person.

## Proposed fix

Separate the two failure modes so each surfaces an accurate message:

```ts
let pipeline;
try {
  ({ pipeline } = await import('@xenova/transformers'));
} catch (cause) {
  throw new Error(
    '@xenova/transformers is not installed. Install with: npm install @xenova/transformers',
    { cause },
  );
}
return (await pipeline('feature-extraction', model, { quantized: true })) as PipelineFunction;
// step (B) errors now propagate with the real cause
```

- Import failure → "not installed" (actionable install hint), with original error as `cause`.
- Pipeline/model failure → propagate the original error (or wrap with context preserving `cause`).

## Notes

- Audit sibling lazy-loaded providers (e.g. `onnx-embedding-provider.ts`) for the same over-broad-catch pattern.

daviddgonzalez:
Closes #118.

## Problem

`loadPipeline` in `packages/rag/src/embedding-providers/transformers-embedding-provider.ts` wrapped **both** the dynamic `import('@xenova/transformers')` and the model download/inference in a single `try/catch` that always rethrew a fixed `@xenova/transformers is not installed` message. Any failure in step B (model download, network error, `onnxruntime-node` native backend, OOM, …) was reported as "not installed" even when the package was installed and importable — and the real error was swallowed, not even attached as `cause`.

This actively misled debugging: a Windows CI failure was initially diagnosed as a missing/optional dependency when the package was in fact installed; the real failure was the model download / native backend. PR #119 fixed the test-reliability symptom by skipping those tests on Windows, but the misleading error remained.

## Fix

**`transformers-embedding-provider.ts`** — split the two failure modes so each surfaces an accurate message:

- Import failure → keeps the actionable `@xenova/transformers is not installed` hint, **now with the original error attached as `cause`** (previously dropped entirely).
- Model/inference failure → throws `Failed to load transformers model '<model>'` preserving the original error as `cause`.

A type-only import of `pipeline` types the holder variable across the two `try` blocks without an `import()` type annotation (lint) or any runtime dependency on the optional package.

**`onnx-embedding-provider.ts`** (the issue's "audit siblings" note) — its install-hint `catch` (`loadOnnxRuntime`) was **already correctly scoped** to the import alone, so it was left unchanged. But `loadModel`'s model download (`ensureModelFiles`) and session creation (`InferenceSession.create`) previously bubbled raw errors with no provider/model context, so they now throw `Failed to download ONNX model '<model>'` / `Failed to load ONNX model '<model>'` with `cause` preserved.

## Tests

No new tests. This is a pure error-message refinement adding no new branches that aren't already exercised by the existing integration paths; the happy path stays covered by the integration tests. Full `bun run validate` (lint, typecheck, unit + integration + system tests, duplication) passes.

## Changelog

Added an entry under `### Fixed` in the `[Unreleased]` section.


daviddgonzalez:
This is just to do with I/O which will be tested in integration tests. Putting it in unit tests is redundant/might confuse other claudes that looks at the test suite

jdutton:
@daviddgonzalez  you can merge this whenever you are ready

PR #137, opened 2026-06-26
feat(cli): add top-level `vat validate` command (#128)

### LINKED ISSUE #128: Add a top-level `vat validate` that runs every configured validator (resources + skills + …) so consumers stop calling partial subcommands
## Summary

There is no top-level `vat validate`. A project must call each validator separately — `vat resources validate` (docs/links) and `vat skills validate` (skills) — and hand-wire them into a composite script. That is leaky and fragile, and it trains people to call partial commands.

## Request

Add `vat validate` that runs **every validator the project's config declares — and only those**. It should inspect the config (`resources`, `skills`, and any future surfaces) and run each configured validator, skipping surfaces the project doesn't use. Aggregate the results; exit non-zero if any fails.

## Why this matters

- **One command to teach and to run.** "Run `vat validate`" is the whole story for local pre-commit and CI. No one has to remember the list of subcommands.
- **No partial-coverage footgun.** A contributor who runs `vat resources validate` today silently skips skill validation and thinks they're clean. That is the core problem: the partial command *looks* like "validate" but isn't. `vat validate` removes the trap.
- **Future-proof by construction.** When a new validator lands (agents, MCP manifests, settings, …), `vat validate` picks it up from config with zero downstream changes. Every consumer who hand-composed `resources validate && skills validate` would otherwise have to discover the new surface and add it — and most won't, so coverage silently rots.
- **"Doesn't check what it doesn't know about."** Because it is config-driven, a project with no `skills:` block simply doesn't run skill validation — no error, no noise. The command validates exactly the declared surfaces.

## Suggested behavior

- `vat validate` → for each configured surface, run its validator in a stable order (e.g. `resources`, then `skills`, then future), print a combined summary, exit `0` only if all pass.
- `vat validate --only <surface>` for parity with `vat build --only <phase>`.
- Functionally equivalent **today** to `vat resources validate && vat skills validate` — but discovered from config instead of hand-composed, so it cannot drift out of coverage.

## Environment

`0.1.39-rc.1` — bare `vat validate` returns `error: unknown command 'validate'`; only `resources validate`, `skills validate`, and `agent validate` exist as `validate` subcommands.

daviddgonzalez:
Closes #128.

## What

Adds `vat validate` — a single top-level command that runs the **source-level** validators the project's config declares, and only those:

1. `resources validate` (when `resources:` is configured)
2. `skills validate` (when `skills:` is configured)

It aggregates phase results to YAML, exits non-zero if any fail, and supports `--only <surface>`. This replaces the hand-composed `vat resources validate && vat skills validate`, which silently loses coverage when someone forgets a subcommand.

## Behavior

- **Config-driven:** a surface whose config block is absent is skipped (no error, no noise) — matching the issue's "doesn't check what it doesn't know about" requirement. A bare run with *nothing* configured at all still exits 0, but now also emits a stderr warning — a config typo (e.g. `recources:`) can no longer produce a silent, indistinguishable-from-passing exit-0 run.
- **Project-root aware:** config is read from the resolved project root, so a run from a subdirectory still discovers the project's surfaces (rather than finding an empty config and falsely passing).
- **`--only` safety, and consistent exit codes:** an explicit `--only <surface>` fails (**exit 1**) whether the named surface is unrecognized or simply not configured — both are "you asked for a surface that can't run," and both now land on the same code instead of splitting across 1 (unconfigured) and 2 (unrecognized, via the generic system-error path). A CI gate differentiating "validation failed" from "system error" can no longer miss one of the two failure modes.
- **Source-level only:** unlike `vat verify`, it never inspects built `dist/` artifacts and never requires a build — safe for pre-commit and CI-before-build.

## Decision (revisitable)

Marketplace-artifact validation is **intentionally excluded**. It runs against the built `dist/` marketplace tree, which would couple `vat validate` to a prior `vat build` and overlap `vat verify`. So the split is:

| Command | Scope | Build needed? |
|---|---|---|
| `vat validate` | source: resources + skills | no |
| `vat verify` | built artifacts: + marketplace + consistency | yes (after `vat build`) |

This is called out in the code and in `docs/guides/marketplace-distribution.md` so it can be revisited if a single "validate the whole shippable thing" command is later wanted (just fold the marketplace phase in, mirroring `verify.ts`).

## Docs

- `docs/guides/marketplace-distribution.md` — split Validation into source (`vat validate`) vs built-artifact (`vat verify`); fixed the publish flow and mode table.
- `docs/architecture/cli.md` — dropped the "Future" label.
- `packages/vat-development-agents/resources/skills/vat-adoption-and-configuration.md` — the CI-wiring `vibe-validate.config.yaml` sample now includes a `vat validate` phase before `vat build`/`vat verify`, matching the surrounding prose (previously the sample only showed build+verify, so an agent copying it verbatim would never adopt the new command).
- `CHANGELOG.md` — Unreleased entry, updated to describe the finalized (hardened) behavior below.

## Tests

`packages/cli/test/system/validate.system.test.ts` — 7 system tests: resources-only, skills-only, subdirectory discovery (regression), no-op note + stderr warning when nothing configured, `--only` unconfigured fails with exit 1, `--only` unrecognized fails with exit 1 (same code as unconfigured, pinned exactly — was previously only asserted `not.toBe(0)`), and a configured validator failing → non-zero exit.

## Review history

1. **Initial workflow-backed code review** (8 finder angles + per-finding verification) ran on the diff. Three findings were fixed (subdirectory false-green, silent `--only` pass, docs overpromising auto-discovery).
2. **Independent adversarial UX review**, run cold after the PR sat for two weeks (treating it as unreviewed), focused on developer/AI-agent user experience rather than just code correctness. It confirmed the core design (separate command, config-gated, source-vs-built split) is sound, but surfaced three real, verifiable issues, all now fixed in this PR:
   - The `--only` exit-code split described above (1 vs 2 for what a caller experiences as the same failure mode).
   - The fully-unconfigured bare-run silent-success gap described above.
   - A stale CI-wiring sample (prose said "use `vat validate` for pre-build checks," the adjacent YAML example didn't include it).

   It also flagged, and this PR fixed, a top-level `--help` description asymmetry: `vat verify`'s one-liner says "run after vat build" but `vat validate`'s didn't symmetrically say "no build required" — an agent skimming `vat --help` had no differentiator at that glance level. Now: `Validate configured surfaces from source (resources + skills) — no build required`.

Full validation (`bun run validate` / `vv validate`) passes — lint, typecheck, duplication check, unit/integration/system tests — after a rebase onto current `main` (65+ commits, including the HTML-resources and linkAuth work landed since this PR was opened).


jdutton:
Thanks @daviddgonzalez — the design here is clean (config-driven, `validate` ⊂ `verify`, `--only` on an unconfigured surface fails closed) and the `docs/` work is genuinely thorough: the `marketplace-distribution.md` rewrite teaches the source-vs-built split well (the 4-step `validate → build → verify → publish` flow and the "two layers" section are exactly right), and `cli.md` is fixed. CI is green across the board.

One concern, scoped to **command legibility** given how close the two names are. The split is taught well in `docs/`, but two *other* surfaces a user or agent will actually hit weren't updated to match:

**1. `vat verify --help` is now the stale one.** Its description still reads:
> "Validates all project artifacts **without building anything**."

That (a) never mentions `vat validate`, and (b) collides on the exact phrase `validate`'s help uses to distinguish itself ("never requires a build") — so a reader comparing the two `--help` outputs sees both claim "no build," which is the opposite of disambiguating. It's also slightly inaccurate: verify's `marketplace` + `consistency` phases read `dist/`, so it depends on a prior build even though it doesn't run one. Suggest tightening verify's description to something like:
> "Verify the built distribution (resources + skills + marketplace + consistency). Inspects `dist/` artifacts — run after `vat build`. For source-only pre-commit checks, see `vat validate`."

**2. The agent-facing skills weren't updated.** None of `packages/vat-development-agents/resources/skills/*.md` mention `vat validate` — they still present `vat verify` as the catch-all gate. Since these skills are the routing surface an agent reads, it'll keep recommending `vat verify` (which assumes a build) for source / pre-commit checks. Specific spots:
- `vat-adoption-and-configuration.md:164-168` — the sample CI step and the "authoritative gate before `npm publish`" text point only at `vat verify`; the CI-before-build step is exactly what `validate` is for.
- `vat-skill-distribution.md:440` — quick-ref table row `Verify everything | vat verify` has no `validate` counterpart.
- `SKILL.md:48` — quick-ref `vat verify  # validate all artifacts`, no `validate` line.

Neither is a blocker on the code — `validate.ts` itself is solid (and correctly inherits the global `--cwd`). Happy to take the skill edits + verify-help tweak as a fast follow if you'd rather keep this PR tight. Just flagging so the two near-identical names don't drift apart in their self-documentation.

PR #164, opened 2026-07-28
fix(skill-lifecycle)!: quality audit — make the eval/packaging boundary mean one thing

### LINKED ISSUE #163: Feature: support eval suites + fixtures stored outside the skill source tree
# Feature: support eval suites + fixtures stored outside the skill source tree

## Summary

`vat skill test run <skill>` can only read its eval suite and fixtures from **inside the skill's own source tree** (`<skill>/evals/`). Provide a supported way to point a run at an **external eval root** (an absolute path outside the repo), so eval suites and their input fixtures can live somewhere other than the committed skill source.

## Motivation

Some skills can't keep their eval material in git alongside the skill:

- **Data sensitivity.** A skill that processes regulated or confidential input (PII, health records, financial documents) needs realistic fixtures to be tested meaningfully, but those fixtures must never be committed to a git repository. Today the only way to run `vat skill test` against them is to place them under `<skill>/evals/`, i.e. inside the working tree — exactly where they shouldn't be.
- **Fixture size.** A skill that operates on large binary inputs (multi-gigabyte media files, large sample datasets, model artifacts) shouldn't have those inputs committed to git. They naturally live on a shared drive, object store mount, or local cache outside the repo.
- **Shared/curated corpora.** Teams often maintain a separate, access-controlled corpus of labeled examples that several skills test against. Duplicating or copying it into each skill's `evals/` dir is wasteful and drifts out of sync.

In all three cases the desired workflow is identical: *keep the skill source in git, keep the eval suite + fixtures at an external path, and run the existing harness against them.*

## Current behavior (as of the current release)

From the skill-test harness:

- The suite is always resolved as `<subjectStagedDir>/<evalsSubpath>`, where `evalsSubpath` defaults to `evals/evals.json` and is staged from the skill's own source dir.
- The only override, the `evals` config key under a skill's `test` block, is documented and validated as a **relative subpath** ("Path to evals.json (relative to skill source)"). It is joined onto the staged skill dir, so an absolute value is concatenated into a non-existent path rather than honored.
- `${fixturesDir}` resolves to `<stagedSkillDir>/<evalsDir>/fixtures` — computed, not configurable.
- An eval entry's `files: [...]` are copied from paths **contained under** the evals dir; absolute segments and `..` are rejected.
- There is no CLI flag and no environment variable to redirect where evals or fixtures are read from. `--workdir` / `--out` only control where the temporary harness is written, and `--workdir` is additionally validated to reject any location under a `CLAUDE.md` / `.claude` tree.

Net: **external eval/fixture roots are not supported** — absolute paths are either silently mangled or hard-rejected.

## Proposed feature

Allow a run to specify an **external evals root** (absolute path), from which the suite and its fixtures are staged instead of `<skill>/evals/`.

Suggested surface (open to design):

- A CLI flag, e.g. `--evals-root <abs-dir>` (containing `evals.json` and a `fixtures/` subdir), or `--evals <abs-path-to-evals.json>`.
- And/or allow the existing `test.evals` config value to be **absolute** (with a clearly documented "outside the skill source" contract), plus a companion `test.evalsRoot` / `test.fixturesDir`.
- The chosen base must be threaded consistently into: the suite read, `stageEvalWorkspaces` (so `files: [...]` resolve under the external root), and the `${fixturesDir}` token computation — so all three come from the same external base rather than the staged subject dir.

### Behavior details

- When an external root is provided, stage the suite + fixtures from there into the temp harness exactly as today (so the executor still runs against a staged copy, not a live external path). Only the **source** of the staged material changes.
- `files: [...]` and fixtures resolve **relative to the external root**, keeping the existing containment checks (no escaping the external root via `..`).
- If no external root is given, behavior is unchanged (staged from `<skill>/evals/`).

## Acceptance criteria

- [ ] A run can be pointed at an absolute external directory whose `evals.json` + `fixtures/` are used for the run, with no eval material required under the skill's git-tracked source.
- [ ] `${fixturesDir}` and eval-entry `files: [...]` resolve against the external root and stage correctly.
- [ ] Containment/safety checks still apply within the external root (no `..` escape).
- [ ] Absolute-path inputs are honored, not silently concatenated onto the staged skill dir.
- [ ] Omitting the option preserves current behavior exactly (default `evals/evals.json`).
- [ ] `--dry-run` reports the resolved external evals/fixtures locations so misconfiguration is caught before a paid run.
- [ ] Documented, including the security note below.

## Security / safety considerations

- The executor already runs staged skill code with elevated permissions; reading eval inputs from a user-specified external path does not broaden that trust boundary, but the docs should state plainly that the external root is operator-controlled and is treated as trusted input material.
- Context isolation should be preserved: the external root is a *fixture/eval source*, not a place to load settings or memory from. The existing "refuse a workdir under a `CLAUDE.md`/`.claude` tree" protection should not be undermined by this feature.
- Because a motivating use case is sensitive data, the harness should avoid copying the external material anywhere durable beyond the temporary harness dir, and should honor the existing `--keep` semantics (i.e. don't leave sensitive fixtures behind unless explicitly asked).

## Alternatives considered

- **Pre-stage build hook** (`test.build`) that copies external evals/fixtures into `<skill>/evals/` before a run. Works today, but for the sensitive-data case it routes the sensitive material *through the git working tree* — the exact thing to avoid — and relies on `.gitignore` discipline to prevent an accidental commit.
- **Keep the whole skill source outside git.** Defeats the purpose of versioning the skill; only the *eval material* needs to be external, not the skill.

## Notes

Discovered while designing a testing workflow for a skill family whose realistic test inputs cannot be committed to git for data-sensitivity reasons. The `files: [...]` + context-isolation features are already most of the way there; the missing piece is decoupling the eval/fixture **source root** from the skill's committed source tree.

jdutton:
> **Scope note — this is a skill-lifecycle quality audit PR.** It started as "fix the eval answer-key leak." That bug was real, but chasing it exposed that the eval/packaging boundary was never specified coherently: each lane had drifted to its own answer about where a skill's test input lives and what that implies. Everything below is what we found tearing that wall open, not what we set out to fix. The findings are grouped by the defect *class* they belong to rather than by the symptom that surfaced them.

## The unifying defect

**A VAT lane reasons about the SOURCE tree while the artifact that ships or gets tested is the BUILT tree.** Nearly every bug here is that shape, and the ones that aren't are its mirror: two lanes answering the same question differently.

## What changed

**Eval answer-key isolation.** The suite (`evals.json` plus its `fixtures/`) no longer reaches the staged tree the executor runs against, by any resolution route. A canary walks cwd, sandbox, plugin dirs and the whole harness root at spawn time.

**`<skill-root>/evals/evals.json` is now a declaration in both lanes.** The harness always auto-detected it; the packager refused to, so a skill following the convention had its answer key **published** while the harness quietly protected the eval signal. Pure auto-detect — keyed on the suite *file*, naming exactly `<skill-root>/evals`. A root `evals/` with no suite file, and any other directory named `evals`, still ship. No suite at all is a clean no-op; evals never become mandatory.

**BREAKING: plugin-local skills are packaged, not tree-copied.** They previously shipped verbatim — eval suites, scratch files, un-rewritten links and all.

**BREAKING: `${fixturesDir}` resolves per eval.** It pointed at a directory this branch deletes, and failed silently because interpolation has no existence check. Now `<eval workspace>/fixtures`, which is also the executor's cwd. An eval with no declared `files` fails preflight rather than receiving a dead path.

**Packaged link fidelity.** Three fixes: a link with no packaged target strips to prose in every spelling; a stripped link keeps *its own* text (an href-keyed "first wins" map made a second link render the first one's words — this shipped in VAT's own `vat-skill-review`); and link syntax inside code blocks is no longer rewritten, so a skill teaching authored syntax stops shipping the packaged path instead.

**`--dry-run` builds, once acknowledged.** A preview assembled from a stale dist previews something a real run would not test. Gated on the raw acknowledgement flag — not on `acknowledged`, which every dry run synthesizes — because building runs the repo's `test.build` shell hook.

## The gate that makes this stick

Two of the worst defects here shipped through a **fully green 14-step `validate`**, because nothing compared built output against a baseline. `packages/vat-development-agents/test/system/packaged-output-drift.system.test.ts` now diffs all 12 packaged `SKILL.md` files against a committed golden tree, so any packager change surfaces as a reviewable diff in the PR that causes it. It was verified by reintroducing a real corruption and confirming it fails with a readable diff — not assumed to work.

## Verification

`bun run validate` green (14 steps, 3 phases). Every fix has a test written first and **verified to fail against the old behavior** — including reverting each fix to confirm the test catches it. Evals run on the touched skill: 24/25 expectations, the single miss a terseness judgment the grader confirms is not a content regression.

## Found, filed, not fixed

- #165 — no detection of a stale `files:` build artifact (`integrity` is intra-build and structurally cannot see it)
- #166 — `test.evals` in a shared directory strips that whole directory
- #167 — a link to a `files:` dest *directory* fails validation

## Discovered: external skill testing is half a feature

`vat skill test` supports external skills (`npm:`, `url:`, `vendored`, undeclared paths) — staged as-is, no build. But eval lookup only searches inside the skill's own tree, and a VAT-built package **provably has no evals**: they're excluded from `dist/`, and the published `files` array omits `resources/` entirely. So the only external skill you can eval is one whose publisher violated the rule VAT enforces. Bootstrap would write a starter suite into `node_modules`.

That reframed #163 from an enhancement into the missing half of a shipped feature. **It is now fixed on this branch** (`849de5ee`): `--evals <path>` accepts a relative path, an absolute path, or an npm bare specifier, resolving against the **current directory** so an operator's shell path means what they typed, while config `test.evals` continues to resolve against the **skill source**. Each eval's declared input `files` resolve relative to the *suite's* directory, so fixtures travel with the suite. A bug fixed in passing: an absolute `test.evals` was folded into `<skillDir>/<path>` by a plain path join, and rather than failing, VAT bootstrapped a starter template at that bogus path and exited 3.

Closes #163.

## Added after this description was written

The audit continued past the boundary above; the CHANGELOG carries the full detail. In lifecycle order:

- **Install / publish lanes name a skill by its declared name, not its path** (`09192558`) — a flat ZIP installed as `vat-skills-install-zip-<random>`, the extraction temp dir. And `vat claude org skills install` had **no exclusions whatsoever** (`ff3345ff`) — point it at a source tree and the answer key published to the whole org.
- **A staged-tree inventory gate** (`0cd66a0a`) — every prior staging test asserted one *named* property, so all of them kept passing when staging emitted an extra file, which is exactly what a leaked answer key is.
- **`vat resources validate` now says what severity each finding is** (`223580b0`, BREAKING) — `error*` fields count errors only, `issue*` fields count all severities; text format gained the severity. A scan could previously report `status: success` beside `filesWithErrors: 1` beside `errorsFound: 0`. Also: unparseable frontmatter is no longer *also* reported as missing frontmatter.
- **`vat audit <path>` never reaches the network for a path that exists** (`7676b78a`, BREAKING for `owner/repo` shorthand) — `vat audit plugins/arc` cloned `https://github.com/plugins/arc.git`. Same commit: audit built the document registry once per skill, twice over, and the inventory walk asked git a question that abandoned `git ls-files` for a full recursive walk (**39,599 ms vs 16 ms** for the same ~1,146 files). Measured after: a plugin 369 s → 42 s; 46 skills did-not-finish-in-10-min → 571 s.
- **Dot-directories are no longer invisible to the file crawler** (`cfa96ac6`, BREAKING — widens the scan). `**` would not traverse a segment beginning with a dot, so `**/*.md` could never see inside `.claude/` — the home of the rules, skills, commands and agents VAT exists to validate. `vat audit` already compiled its `--exclude` patterns with `dot: true`, with a comment saying that without it the exclude "silently never fires"; the include side never got the same treatment. One question, two lanes.
- **Explicit `<a id="short">` anchors in markdown resolve** (`6d5c3303`) — VAT indexed heading slugs only, so a short hand-written anchor above a long heading reported `LINK_BROKEN_ANCHOR` for a link GitHub resolves. Only mdast `html` nodes are read, so an `id=` shown inside a code fence is *documented*, not declared, and is never indexed.

- **`ValidationIssue.location` means one thing** (`1231f8ec`, BREAKING). It was carrying six incompatible kinds of value at once: absolute paths, absolute paths with `:line` glued on, relative paths (the resources lane's convention leaking through), document-internal pointers that are not paths at all (`frontmatter.description`), Zod field paths joined by the **same** colon a line number uses, and — for walker-derived link findings — the link **target** rather than the file containing the link. So a consumer could not parse it uniformly, could not know what to resolve it against, and absolute paths leaked the developer's home directory into CI logs. Now four fields, one meaning each: `location` (the file you would open, project-relative POSIX, **never** absolute), `line` (a number — the field already existed and no skills-lane producer had ever set it), **`field`** (new: a dotted pointer *inside* that document), and `link` (a link href/target, never the file to open).
  - The gate came first and is the point: a contract test drives both skills-lane entry points over one fixture and asserts the invariant over **every** issue emitted, not a named subset — so a new producer with a new convention is caught the day it lands. All three defect classes were reversion-proved. `location` also gained a schema refinement; it was a bare `z.string()`, which is how 235 absolute paths shipped unnoticed.
  - **Adopter action required:** `validation.allow` globs for link codes now match the issue's `location` **or** its `link`. Entries written against the resolved *target* no longer match — rewrite them against the containing file, which is usually shorter and more precise (VAT's own six-pattern entry collapsed to one). Target globs were depth-fragile regardless: picomatch's `**` does not cross a leading `../`, so each nesting level needed its own `../../../…` pattern. VAT's own dogfood build caught this by going red.
  - The frozen audit-parity snapshot was **re-anchored, not redefined** — gated on proving the new run reproduces all 200 legacy tuples exactly on `(path, code, severity)`, the dimensions the change does not touch, with zero losses.
- **Plugin manifest merge warnings name the winning value and where it came from** (`1231f8ec`). An adopter reported VAT "silently overwriting" their authored `plugin.json` `author` on every build. The behaviour is correct — `author` is the marketplace `owner` from their own `vibe-agent-toolkit.config.yaml`, exactly like `name` and `version` — but the warning called it "VAT-generated" and named neither the winning value nor a remediation, while the sibling version-mismatch warning already said "Reconcile by removing one or the other". Same shape as the severity fix above: the policy was right, the report was the only thing that said otherwise.

## What this audit implies for the next one

Every defect here was found by asking one question — *does this lane agree with the others about what a skill's test input is?* — and following it past the symptom that prompted it. The answer was no, in five distinct places, three of which were shipping wrong bytes to users.

That question was scoped to one boundary. The same treatment is owed to the **whole skill lifecycle**: author → validate → build → package → publish → install → test. The standard is the one applied to security audits, not to code review: *does this source set out to do a coherent thing, and does it actually accomplish it?* Not "is this line correct" but "is there a single intended contract here at all, and does every lane implement the same one?"

Findings from that audit become backlog, and the backlog gets fixed — the same way this PR turned one reported bug into five.


jdutton:
Superseded by #169, which now carries this branch's full scope.

#169 merged this branch (91 commits) plus the two commits that were still unpushed here (`2365ada6` retiring `vat validate --only`, and its changelog), and the three uncommitted doc corrections in that worktree — including the one recording that a filename collision is *reported, not thrown*, and must not be used as the fixture for the batch-containment guard.

A no-loss audit over all 442 changed files confirms nothing was dropped: of the 416 files belonging to this branch alone, 401 are byte-identical in #169, 2 are this branch's own intentional deletions, and 13 differ only by work added on top.

The cross-skill answer-key leak flagged here is fixed in #169 and is wired through every lane. Closing in favour of one PR.

PR #188, opened 2026-08-07
fix(packaging,audit): degrade on non-regular glob matches and unreadable scan paths (#183, #180)

### LINKED ISSUE #180: vat audit: one unreadable subdirectory aborts the whole run (exit 2) and loses every finding
## Summary

One unreadable subdirectory anywhere under an audited tree aborts the **entire** `vat audit` run:
`status: error`, exit code 2, and **zero findings** — including the findings already collected from
readable siblings. A single root-owned or quarantined directory under `~/.claude/plugins` kills the
flagship `vat audit --user` invocation outright.

This is **pre-existing and out of scope for #170** — the crawl call site is unguarded independently
of anything that PR changed. It is filed here because #170's adversarial review is what surfaced it.

## Reproduction

Input tree — a bundle with a readable `CLAUDE.md` at its root and a second one in a subdirectory
that is then made unreadable:

```
<project>/skills/demo/SKILL.md
<project>/skills/demo/CLAUDE.md
<project>/skills/demo/sub/CLAUDE.md
```

```bash
vat audit <project>/skills/demo          # readable: baseline
chmod 000 <project>/skills/demo/sub
vat audit <project>/skills/demo          # same command, one directory unreadable
```

Observed:

```
readable:    2 findings, exit 0

chmod 000:   Audit failed: EACCES: permission denied, scandir '.../skills/demo/sub'
             status: error
             exit code 2
             0 findings   <- the readable root-level CLAUDE.md is lost too
```

Expected: the readable root-level `CLAUDE.md` is still reported, plus one finding (or one `info`)
naming `sub/` as unscannable. An audit over a tree it does not own must not lose the work it already
did because one entry is inaccessible.

## Cause

`detectPackagedAgentInstructionFiles` calls `crawlDirectorySync` with no guard —
`packages/agent-skills/src/validators/agent-instruction-presence.ts` (the `crawlDirectorySync({...})`
call, around line 70):

```ts
const files = crawlDirectorySync({
  baseDir: rootDir,
  include: INCLUDE_GLOBS,
  exclude: [],
  absolute: true,
  filesOnly: true,
  respectGitignore: false,
});
```

The `EACCES` throw propagates out of `distributedTreeFindings`
(`packages/cli/src/commands/audit/distributed-tree.ts`, the
`return detectPackagedAgentInstructionFiles(...)` tail) straight to the top of the command, so no
partial result survives.

This contradicts the stance `packages/cli/src/commands/audit.ts` takes everywhere else — audit is a
bulk linter over trees it does not own, and a broken governing config must not abort the scan.

## Suggested fix

Wrap the crawl in `detectPackagedAgentInstructionFiles`:

```ts
let files: string[];
try {
  files = crawlDirectorySync({ ... });
} catch (err) {
  return [materializeIssue('...UNSCANNABLE_TREE', { /* names the unreadable path */ })];
}
```

If a new code is unwanted, catch in `distributedTreeFindings` instead and emit one `info` naming the
unreadable path. Swallowing silently is **not** acceptable — a scan that reports success while
having skipped a subtree is the same failure shape as a detector that silently disables itself.

Ideally the guard is per-directory rather than per-tree, so an unreadable leaf costs one finding
rather than the whole subtree.

## Severity

High — directly reachable by `vat audit --user`, the flagship invocation, and the failure mode
destroys every finding the run had already produced rather than degrading. Found during the
adversarial review of #170.

### LINKED ISSUE #183: A files: glob matching a symlink-to-directory kills the build with an unattributed ENOTSUP
## Summary

A `files:` glob that matches a **symlink pointing at a directory** kills the build with a raw OS
error. `nodir: true` does not exclude it — `glob` does not follow the link, so it is never
classified as a directory — and the copy loop calls `copyFile` on it, which throws `ENOTSUP`. The
message names neither the `files:` entry that caused it nor any remediation, and on macOS the
`ENOTSUP` string ("operation not supported on socket") actively misdescribes the object.

This is **pre-existing and out of scope for #170** — the copy loop's missing `lstat` guard predates
that branch. Filed from #170's adversarial review.

## Reproduction

Input tree — `assets/linkdir` is a symlink to the directory `assets/sub`:

```
assets/sub/               (a directory)
assets/linkdir -> sub     (symlink to a directory)
assets/ok.mjs
```

Config:

```yaml
version: 1
skills:
  include: ["resources/skills/SKILL.md"]
  defaults:
    targets: ['claude-code']
  config:
    demo:
      files:
        - source: "assets/*"
          dest: scripts
```

`vat skills build` →

```
Build failed for skill: demo
   ENOTSUP: operation not supported on socket, copyfile '.../assets/linkdir' -> '.../scripts/linkdir'
   Nothing was written — dist/skills does not exist
Build failed: 1 skill(s) could not be packaged at all
```

Expected: `assets/linkdir` is skipped (or resolved deliberately), with a finding that names
`source: "assets/*"` and the path, and the rest of the entry ships.

Controls, both verified:

- A plain **directory** *is* correctly excluded by `nodir: true` — a directory named
  `assets/adir.mjs` never appeared in the output.
- A symlink to a **file** is copied by content, which is reasonable behaviour and should stay.

So the gap is specific to symlink-to-directory (and, by the same mechanism, FIFOs, sockets and
device nodes).

## Cause

`packages/agent-skills/src/files-config.ts`:

- `expandGlobEntry` (around line 496) runs
  `glob(remainder, { cwd: absoluteBase, nodir: true, dot: true })`. `nodir` filters on what `glob`
  itself stat'd; an unfollowed symlink is not a dir to it.
- The copy loop (around lines 579-594) calls `await copyFile(absSource, absDest)` on every match with
  no type check, so the raw `ENOTSUP` escapes as the build's whole explanation.

## Suggested fix

In the copy loop, `lstatSync` each match and skip anything that is not a regular file, raising a
`FILES_GLOB_*` finding that names `entry.source` and the offending path. The `lstat` guard is
preferred over passing `follow: false` plus `stat: true` to `glob` (which would resolve symlink
types before `nodir` applies) because `lstat` also covers FIFOs, sockets and device nodes with one
predicate.

Whichever route, the error must name the `files:` entry — a build that dies on an OS errno with no
attribution gives the author nothing to act on.

## Severity

Medium — it is a hard build failure rather than a silent wrong result, but it is unattributed and
the OS text is actively misleading, so the diagnosis cost is high relative to the trivial cause.
Found during the adversarial review of #170.

ejdutton:
Two pre-existing failures where a single bad filesystem entry destroyed the whole run's output instead of costing one finding. Both were reproduced before the fix and re-verified end-to-end through the real CLI after.

Closes #183
Closes #180

---

## #183 — a `files:` glob matching a symlink-to-directory killed the build

`glob`'s `nodir: true` cannot exclude it: glob does not follow symlinks, so a link pointing at a directory is never a directory *to it* and arrives as an ordinary match. `copyFile` then threw a raw `ENOTSUP` that escaped as the build's entire explanation, naming neither the `files:` entry nor the path — and on macOS it renders as *"operation not supported on socket"*, misdescribing the object outright.

Reproduced first, verbatim:

```
Error: ENOTSUP: operation not supported on socket, copyfile '.../assets/linkdir' -> '.../scripts/linkdir'
 ❯ copyGlobEntry src/files-config.ts:612:5
```

**Fix.** Non-regular matches are partitioned out inside `expandGlobEntry` — the *one* expansion the copy and the pre-build gates share, so the lanes cannot disagree about what ships. They are reported as the new `FILES_GLOB_SKIPPED_NON_REGULAR_FILE` (`warning`), and the rest of the entry ships.

**A symlink to a regular FILE is still copied by content.** The predicate is "does this resolve to a regular file, without throwing" — not "is this a symlink" — which keeps that legitimate, widely-used case working while also covering dangling links, FIFOs, sockets and device nodes without enumerating them. An `isSymbolicLink()` test would have broken the good case; a bare `stat().isFile()` would have put the same unattributed errno back on the critical path via dangling links.

After, through the real `vat skills build`:

```
status: warning        skillsBuilt: 1        errors: 0
code: FILES_GLOB_SKIPPED_NON_REGULAR_FILE

dist/skills/demo/SKILL.md
dist/skills/demo/scripts/ok.mjs     <- ships; linkdir correctly absent
```

`vat skills validate` reports the same finding pre-build, anchored at `assets/linkdir` and naming `files: source 'assets/*'`.

## #180 — one unreadable subdirectory aborted the entire `vat audit` run

`status: error`, exit 2, and **zero findings** — discarding everything already collected from readable siblings. One root-owned or quarantined directory under `~/.claude/plugins` killed the flagship `vat audit --user` outright.

> [!IMPORTANT]
> **The cause recorded in #180 is not the one.** The issue points at `crawlDirectorySync` in `agent-instruction-presence.ts`. That crawl already guards each directory individually (`walkDirectory` catches and returns). Traced with an `fs` interceptor, the unguarded `readdir` is `scanDirectory`'s own, in `commands/audit.ts`. Fixing where the issue said to would have changed nothing.

**Fix.** The failure is scoped to the subtree that caused it — the walk returns a finding instead of throwing, so the caller's loop keeps its readable siblings — and reported as the new `SCAN_PATH_UNREADABLE` (`warning`).

Silence was rejected deliberately: a scan reporting `success` while having skipped a subtree is the same failure shape as a detector that quietly disables itself. Same stance as `TREE_PROVENANCE_INDETERMINATE` — a missing answer is reported as a missing answer.

Before / after on the issue's own repro:

| | before | after |
|---|---|---|
| status | `error` | `warning` |
| exit | 2 | 0 |
| findings | **0** | root `CLAUDE.md` + `SCAN_PATH_UNREADABLE` at `sub` |

## Severity choices

Both codes are `warning`, within the bar in `docs/validation-rule-design.md`: deterministic filesystem facts rather than smell rules, individually suppressible, and describing a *partial* result rather than a failed one. Neither is an `error`, so neither newly fails anyone's CI.

## Testing

- 6 new cases across `files-config.test.ts` and a new `audit-unreadable-path.integration.test.ts`.
- **Every new test was confirmed to fail against the unfixed code** before being kept — 4 red for #180, 4 for #183 — so none of them pass vacuously.
- The #180 suite skips on Windows and refuses to run as root, where `chmod 000` denies nothing and the fixture would silently assert success.
- The mixed case (a glob netting both policy-refused files *and* non-files) is pinned by its own test: that throw aborts the build, so it is the last word on the entry and must not name only one of the two causes.
- Full `bun run validate` green, plus `vat audit --user` dogfooded over 99 installed skills.

## Note on the CHANGELOG

`FILES_GLOB_SKIPPED_NON_REGULAR_FILE` is folded into the existing **Added** entry rather than getting its own **Fixed** entry. Per `.claude/rules/changelog-adopter-visible.md` the baseline is the last *stable* release: at `v0.1.38` `files-config.ts` had no copy logic at all, so the `ENOTSUP` bug never shipped in a stable version. #180 does get a Fixed entry — its unguarded `readdir` is present at `v0.1.38`.


jdutton:
@ejdutton — adversarial review below. The engineering quality here is high: the comments explain *why* rather than *what*, the symlink-to-file case is exactly the right positive control to have, and the test suite having a readable baseline plus a root-refusal guard is better discipline than most of what lands in this repo. What follows is only what survived me trying to break it.

---

## 1. HIGH — #180 is half fixed: an unreadable **file** still destroys the whole run

The guard wraps `fs.readdir` only. `handleFileEntry` → `validateSkill`/`validate` has no equivalent, so an unreadable *regular file* still reproduces #180's exact signature. Reproduced against a tree where `audit.ts` is unchanged by this PR (the diff here adds only the `Dirent`/`materializeIssue` imports, `unreadableDirectoryResult()`, and the `readdir` try/catch — nothing on this path):

```
# fixture: two skills, one readable, one about to be locked
probe/skills/good/SKILL.md      (valid frontmatter)
probe/skills/locked/SKILL.md    (valid frontmatter)

$ vat audit probe -v
status: success    filesScanned: 2    filesPassed: 2

$ chmod 000 probe/skills/locked/SKILL.md
$ vat audit probe -v
AgentAudit failed: EACCES: permission denied, open '.../probe/skills/locked/SKILL.md'
status: error
exit 2
# zero findings — the readable sibling's result is discarded
```

That is the precise loss this PR set out to stop. Worth noting that the framing in the issue and in your own comment is *ownership/permissions* ("one root-owned or quarantined directory"), while the fix is keyed on *directory-ness*. In `~/.claude/plugins` — where things arrive via `sudo` installs and quarantine — a root-owned `SKILL.md` or `plugin.json` is at least as likely as a root-owned directory. This is the one I'd block on: the headline claim ("degrading beats destroying") currently holds for only half the population.

## 2. MEDIUM — the pre-build gate has no verdict for "matched only non-regular files", and the warning it emits says the opposite of what happens

`copyGlobEntry` **throws** for that case (the new `matches.length === 0 && droppedRel.length === 0` branch). `collectPreBuildGlobFindings` pushes `skipped` warnings and `continue`s with no entry-level verdict. So `vat validate` reports `FILES_GLOB_SKIPPED_NON_REGULAR_FILE` (warning) — *"it cannot be packaged and was skipped, so it did not ship"* — which reads as "the rest of the entry shipped fine". Then `vat build` dies on that entry.

The repo already has the right convention and this is the one place it breaks: `FILES_GLOB_MATCHED_ONLY_NEVER_PACKAGED` is *also* `warning` severity, but its description says outright *"so the entry ships nothing and `vat skills build` fails on it."* The severity isn't the problem — the description is. The two lanes also hand out different remediation for the same state. Cleanest fix is a fourth entry-level verdict mirroring `MATCHED_ONLY_NEVER_PACKAGED`; the minimum is amending the description so it doesn't promise a partial success that isn't coming.

## 3. MEDIUM — both branches from #2 are untested

Grepping the test diff for the new throw text (`none of them is a regular file`) and for an only-non-regular fixture both return nothing. The **mixed** case is tested; the two pure cases are not:

- `copyGlobEntry`'s new "none of them is a regular file" throw
- `collectPreBuildGlobFindings`'s new `kept === 0 && dropped === 0` continue

Also unexercised: dangling symlink, FIFO, socket, device node — all four are enumerated in the registry description *and* in the `isCopyableFile` JSDoc, and none is tested. Dangling is one `symlinkSync('nope', p)`; a FIFO is one `mkfifo` call. Given the mutation-testing posture around here, these are exactly the branches a surviving mutant would name.

## 4. MEDIUM — an unreadable *regular* file in a glob still reaches `copyFile` and throws the raw errno

`isCopyableFile` does `lstat` → `isFile()` → `true`, because `lstat` needs no read permission on the file itself, only on the parent directory. `copyFile` then throws EACCES, unattributed, killing the build — the exact failure mode #183 set out to eliminate, just arrived at through permissions instead of through shape. The predicate answers "what shape is this", not "can I actually read it". Same root theme as #1 above: both halves of this PR fix the shape problem and leave the permission problem.

## 5. LOW — `partitionRegularFiles` adds a serial `lstat` per glob match

Sequential `await` in a `for` loop, one round-trip per match, across every match of every glob of every skill, on the build's critical path; the prior code did zero stat calls. `Promise.all` with a small concurrency cap makes it effectively free.

## 6. INFO — the advertised `severity.SCAN_PATH_UNREADABLE: ignore` hatch can't be set in the scenario it cites

`applySeverityFilter` early-returns unless `deriveConfigRoot(scanPath)` finds a VAT config with a `skills` section. For `vat audit --user` over `~/.claude/plugins` — the case both the issue and your comment name — there is normally no VAT project above that path, so the override has nowhere to live. The *other* remedy you offer, `--exclude`, does work: exclusion is applied per-entry in the parent's loop before descending, so the directory is never `readdir`'d. This limitation is pre-existing and shared with `TREE_PROVENANCE_INDETERMINATE`; flagging only because the new code's motivating example is the one place it doesn't reach.

---

## Checked and clean — don't spend time here

- **The Zod change is safe.** No v3/v4 hazard: `zod` is a direct `dependencies: "^3.23.8"` of `agent-schema`, so `z.ZodTypeDef` resolves against its own v3. And `Partial<Record<IssueCode, …>>` is *more* honest than what `z.record` inferred — the old total `Record` was a lie about a schema that never required all keys.
- `'unknown'` is a legal `ValidationResult['type']`, and status resolves to `warning`, not `error`. Correct.
- **No double-reporting.** `collectPreBuildGlobFindings` (validate lane, via `packaging-validator`) and `packageSkill` (build lane) are reached by different commands.
- The `skipped` / `dropped` split is right — the causes and the remedies genuinely differ, and collapsing them would misinstruct half the time.

## Not yours, but it's blocking this PR

OSV-Scanner is red here **and** on #191, for the same reason: `nanoid` 3.3.16, GHSA-2v37-7h3g-55p8, CVSS 8.2, fixed in 3.3.17. Dependency Audit was green on `main` at `923e19dd` (2026-08-07 14:11) and this PR's run failed at 21:51 the same day — a newly published advisory, caused by neither branch, and it will turn `main` red on the next push too. Nothing for you to do on this branch; flagging so you don't go hunting it.

ejdutton:
Both blocking findings confirmed and fixed. I reproduced each one against the real CLI before touching anything rather than taking the report on faith — and both reproduced exactly as described.

## 1. Unreadable **file** still aborted the whole run

You were right, and this was the more serious of the two: the guard wrapped `readdir` in `scanDirectory` only, so `handleFileEntry` → `validateSkill` threw straight past it.

```
$ chmod 000 skills/bad/SKILL.md && vat audit skills/
AgentAudit failed: EACCES: permission denied, open .../skills/bad/SKILL.md
status: error          # readable sibling's findings destroyed
```

Your point about *why* it matters is the part I had missed: `~/.claude/plugins` is populated by sudo installs and macOS quarantine, so a root-owned **file** is at least as likely as a root-owned directory. Guarding half of it left the motivating scenario broken while looking fixed.

The guard is now per **entry**, covering both branches of the walk. Same input now:

```
status: warning
  SCAN_PATH_UNREADABLE          (bad/SKILL.md)
  PACKAGED_AGENT_INSTRUCTION_FILE   # the sibling finding survives
Audit passed with warnings: 2 file(s)
```

One deliberate refinement on top of your suggestion: **only filesystem access errnos degrade** (`EACCES`/`EPERM`/`ENOENT`/`ELOOP`/…); anything else is rethrown. A blanket `catch` would convert a genuine VAT bug into a `warning` about whichever file it happened on — the same "detector silently disables itself" shape this guard exists to prevent, and it would make the tool quietest exactly when it is most wrong. `ENOENT` is in the set deliberately: a bulk scan races real filesystems, and an entry listed by `readdir` can be gone by the time it is opened.

## 2. `isCopyableFile` and the unreadable-but-regular file

Also confirmed:

```
Build failed for skill: demo
   EACCES: permission denied, copyfile .../assets/locked.mjs -> ...
```

— the bare errno, naming neither the entry nor a remedy. Exactly the defect #183 set out to kill, reached through permissions instead of through file type.

Fixed at the **copy**, not by adding a readability pre-check to `isCopyableFile`. A permission can change between the check and the copy, so a pre-check would only move the race; catching at the copy is correct regardless of cause. Now:

```
files: source 'assets/*' (glob) matched assets/locked.mjs, but it could not be copied
into the bundle: EACCES: permission denied, copyfile ... . Check the file's permissions
and ownership, or narrow the glob so it stops matching this path.
```

It stays a **hard failure** rather than degrading to a skip, and that asymmetry with non-regular matches is intentional: an unreadable file is one the author *declared* and expects in the bundle, so shipping silently without it would change the artifact behind their back. A non-regular match can never be packaged at all, which is why that one degrades. Both are now documented as such.

## 3. The untested throw branch

Covered. The test asserts it raises its *own* error and specifically does **not** reuse the never-package message — whose remedy ("declare an explicit `source:` entry") would not work if followed, since an explicit entry cannot package a directory link either.

## 4. `severity.SCAN_PATH_UNREADABLE: ignore` being unreachable under `--user`

Good catch, and it was my own new text advertising it. The fix string now leads with `--exclude` and qualifies the override as applying only when the scan runs inside a project whose config VAT can find. I did not touch the shared pattern in `TREE_PROVENANCE_INDETERMINATE` — that is pre-existing and belongs in its own change.

## 5. OSV / stale `nanoid`

Merged `main` in (clean auto-merge). That also caught something worth flagging: **0.1.42 shipped in the meantime**, and the merge silently placed all of this PR's changelog entries inside the released `[0.1.42]` section. Moved back under `[Unreleased]`. It also changes the baseline — `copyGlobEntry` exists at `v0.1.42`, so the `ENOTSUP` bug *did* ship in a stable release and now correctly gets a **Fixed** entry rather than being folded into **Added** (which was right when the baseline was v0.1.38).

## Not addressed

`partitionRegularFiles` is still a serial `lstat` per glob match. Agreed it should be bounded-concurrent, and agreed it is not a blocker — leaving it out to keep this diff reviewable rather than widening an already-contested PR. Happy to do it here if you would rather it landed together.

Both new behavioural tests were confirmed to fail against the unfixed code before being kept.

ejdutton:
Fourth review round. Three reviewers with different lenses, plus a duplication check. **No blocking issues remained**, but four findings were worth acting on and are now fixed — two of them because they were claims *in this PR* that were not true.

## Fixed in this round

**An unreadable bundled skill discarded the whole plugin.** 4 results → 1, losing the plugin's own manifest findings and a readable sibling skill, and anchoring the finding at the plugin root, which was perfectly readable. This contradicted both the docs (*"a skill beside the unreadable path is validated exactly as if it were not there"*) and this PR's own code comment (*"costs its own subject and nothing else"*). Now guarded per skill; verified the siblings and manifest findings survive, anchored at `skills/bad/SKILL.md`.

**The CHANGELOG over-claimed the attribution scope.** It said "both glob and explicit entries" — but `claude.marketplaces.<mp>.plugins[].files` is a separate copier and is not covered. Now stated explicitly rather than left to imply coverage it does not have.

**An unreadable config was silently ignored.** The guard added earlier in this PR downgraded a hard failure to `logger.debug`, so a malformed config left every `validation.severity` override inert with no signal at all. That trades a crash for a config that quietly stopped working — the harder problem to notice. Now `logger.warn`, with its own changelog entry.

**The integrity lane said "could not be copied"** when the copy succeeded and *verification* failed, and anchored at an arbitrary first file while the errno named a different one. Both corrected.

## Confirmed working

The reviewer specifically checked the risk I was least sure of: a genuine **content mismatch is not** rewritten as a permissions problem. It throws without an errno, so attribution passes it through untouched.

Also verified clean: `isFilesystemAccessError` cycle-safety and cause-chain depth (real chains are 3 deep, and `ConfigLoadError` preserves `cause`, so the fix reaches the cached lane); `isEmptiedUnreadablePathResult` cannot drop a legitimate result (the only other `type: 'unknown'` producer always carries a `UNKNOWN_FORMAT` issue, and that code is not in `CODE_REGISTRY`, so it cannot be ignored away); marketplace recursion is correctly scoped — an unreadable `pluginB` leaves every `pluginA` finding intact.

## Duplication

**3 clones (0.07%), exactly at baseline — none new.** This was the property genuinely at risk: "fix the class" done badly means copy-pasting parallel guards into each lane, which would have shown up here. The guards consolidated instead.

## Deliberately NOT fixed, named rather than implied away

1. **The plugin build lane** (`plugin-files.ts`, `tree-copy.ts`) has its own raw `copyFile`/`existsSync` and reproduces the same `ENOTSUP` string. It is real and a reviewer grepping `copyFile` will find it — but it is a **different command** from the two this PR fixes, and this branch has already grown across four rounds. Widening it again is precisely how the earlier misses happened. Worth its own issue.
2. **The pre-build gates under-report** an entry whose every match is a non-file: accurate per match, silent about the entry shipping nothing and the build failing. Closing it needs a validation code of its own with the severity justification this repo requires — a design call for the maintainer, not something to slip into round four.

Both are recorded in `docs/validation-codes.md` / `CHANGELOG.md` so neither is discovered by surprise.

## Gate

`bun run validate` — exit 0, zero failures: unit 40.8s, integration 40.6s, system 140s. Every behavioural fix in this PR was verified against the real CLI, and every new behavioural test was confirmed to fail against the unfixed code first.

ejdutton:
Both suggestions taken, and following your repro turned up a correction worth flagging.

## Your repro does not fire in `copyAndRewriteFile`

It is a genuine unguarded site and it is now fixed — but with it guarded, your exact repro still produced the identical bare errno. Traced with an `fs` interceptor:

```
>>> ASYNC readFile(.../skill/scripts/locked.mjs)
    at async parseMarkdown         (resources/link-parser.js)
    at async ResourceRegistry.addResource
    at async registerBundledAssets (agent-skills/skill-packager.js)
    at async packageSkill
```

The throw comes from `registerBundledAssets` → `addResource` → `parseMarkdown` → `readFile` — the step that **reads** every linked file to discover its links, which runs *before* any copier. So the site you named was real, but a different one was on the path your repro takes. Both are guarded now.

Same input today:

```
linked file skill/scripts/locked.mjs, but it could not be read while collecting
the files this skill links to: EACCES: permission denied, open ... .
Check the file's permissions and ownership, and that every directory above it is
traversable.
```

Project-relative path, names the step that actually failed, and a remedy that applies to it.

## Suggestion 1 — one home for the format

Done. `withFsAttribution` is now its own module (`fs-attribution.ts`). `files-config`'s `attributed()` is a thin wrapper supplying the `files:` phrasing; the packager's lanes supply their own subject. A new copier inherits the format by calling it rather than by remembering to reproduce it — which was your point.

One thing added on top: the **remedy is a separate parameter from the action**, because they do not track each other. A failed *read* has nothing to do with whether the output directory is writable or the disk is full, and padding a message with checks that cannot apply is how people learn to stop reading messages.

## Suggestion 2 — the grep pass

36 `await copyFile(` / `mkdir(` / `readFile(` sites across `agent-skills/src` and `cli/src`. Every one accounted for:

| | |
|---|---|
| **Guarded** | both `files:` copy lanes · integrity re-read · skill output-directory `mkdir` (fails before any copier runs) · `copyAndRewriteFile` · `registerBundledAssets` |
| **Tolerant by design** — verified, not assumed | collection-schema read: swallows deliberately so the validator reports instead |
| **Deferred, named** | `vat build` plugin lane (`plugin-files.ts`, `tree-copy.ts`) · `vat skill test` staging lane — different commands |

You were right that this is the only way to know rather than believe. Three of the guarded entries above were found *by* the pass, not by the review.

## One more thing the merge exposed

This branch had drifted behind main, and merging surfaced something in the same family as everything else here: **main converted the `utils` barrel from `export *` to explicit re-exports**, which silently dropped this branch's new export from the `.` entry. The build caught it; the barrel guard test did **not** — it pins the runtime names it is handed, so an export that never arrives is invisible to it. Its docblock says an addition is "fine to accept", but it cannot distinguish an addition you made from one you meant to make. Not introduced here and not fixed here, but worth someone knowing.

## Gate

`bun run validate` exit 0 on the merged tree — unit 42.9s, integration 65.2s, system 178.5s. Duplication 3 clones at baseline, none new; consolidating to one helper could as easily have multiplied the guards, and that is the outcome no test would have shown.

